PR #152, opened 2026-03-20
fix(ambient): remove allowed-tools restriction so skills actually load

dean0x:
## Summary

- **Root cause**: `ambient-router` skill had `allowed-tools: Read, Grep, Glob` which prevented Claude from invoking the `Skill` tool — skills were classified but never loaded
- Removed `allowed-tools` from ambient-router (unrestricted access as main-session orchestrator)
- Added override note in Step 4 `<IMPORTANT>` block so loaded skills' `allowed-tools` don't restrict main session capabilities
- Strengthened hook preamble with explicit Skill tool mention for GUIDED/ORCHESTRATED
- Added `hasSkillLoading()` and `extractLoadedSkills()` integration test helpers + 2 verification tests

## Test plan

- [x] `npm run build` — 35 skills distributed across 17 plugins
- [x] `npm test` — 250/250 pass
- [ ] `npm run test:integration` — verify GUIDED/ORCHESTRATED prompts show "Loading:" marker (requires claude CLI)
- [ ] Manual: new session → GUIDED prompt → confirm Skill tool invocations in output

### DIFF
--- a/src/cli/commands/uninstall.ts
+++ b/src/cli/commands/uninstall.ts
@@ -2,7 +2,7 @@ import { Command } from 'commander';
 import { promises as fs } from 'fs';
 import * as path from 'path';
 import { fileURLToPath } from 'url';
-import { execSync } from 'child_process';
+import { execFileSync } from 'child_process';
 import * as p from '@clack/prompts';
 import color from 'picocolors';
 import { getInstallationPaths, getClaudeDirectory, getDevFlowDirectory, getManagedSettingsPath } from '../utils/paths.js';
@@ -84,7 +84,7 @@ export function formatDryRunPlan(
 function uninstallPluginViaCli(scope: 'user' | 'local'): boolean {
   try {
     const cliScope = scope === 'local' ? 'project' : 'user';
-    execSync(`claude plugin uninstall devflow --scope ${cliScope}`, { stdio: 'inherit' });
+    execFileSync('claude', ['plugin', 'uninstall', 'devflow', '--scope', cliScope], { stdio: 'inherit' });
     return true;
   } catch {
     return false;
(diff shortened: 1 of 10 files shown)


tests/integration/helpers.ts
```diff
@@ -14,16 +15,28 @@ export function isClaudeAvailable(): boolean {
   }
 }
 
+const AMBIENT_PREAMBLE =
```
dean0x: **Issue: Duplicated Magic String** (82% confidence)

The AMBIENT_PREAMBLE is defined identically here and in `scripts/hooks/ambient-prompt:42`. These can drift if one is updated without the other.

**Fix**: Add a test that verifies the preamble substring exists in the hook script:
```typescript
it('preamble matches ambient-prompt hook script', () => {
  const hookScript = readFileSync(
    path.join(__dirname, '../../scripts/hooks/ambient-prompt'),
    'utf-8',
  );
  expect(hookScript).toContain(AMBIENT_PREAMBLE);
});
```

---
_Claude Code Review_

src/cli/commands/uninstall.ts
```diff
@@ -160,6 +190,33 @@ export const uninstallCommand = new Command('uninstall')
...
+      const assets = isSelectiveUninstall
+        ? computeAssetsToRemove(selectedPlugins, DEVFLOW_PLUGINS)
+        : computeAssetsToRemove(DEVFLOW_PLUGINS, DEVFLOW_PLUGINS);
+
+      // Detect extras that would be cleaned up (full uninstall only)
+      const extras: string[] = [];
+      if (!isSelectiveUninstall) {
+        const docsDir = path.join(process.cwd(), '.docs');
+        const memoryDir = path.join(process.cwd(), '.memory');
+        try { await fs.access(docsDir); extras.push('.docs/'); } catch { /* noop */ }
```
dean0x: **Issue: Sequential I/O Performance** (82% confidence)

The two `fs.access` checks for `.docs/` and `.memory/` are awaited sequentially. These are independent filesystem operations that could run in parallel.

**Fix**: Use `Promise.allSettled`:
```typescript
const [docsResult, memoryResult] = await Promise.allSettled([
  fs.access(docsDir),
  fs.access(memoryDir),
]);
if (docsResult.status === 'fulfilled') extras.push('.docs/');
if (memoryResult.status === 'fulfilled') extras.push('.memory/');
```

---
_Claude Code Review_

PR #244, opened 2026-06-16
feat!: close 7 feature-toggle gaps (user-scope-only, clean-break major version)

dean0x:
## Summary

Closes all 7 gaps from the feature-toggle audit where "disabled" didn't fully disable (or there was no off switch). One PR, clean-break **major version**, **user-scope-only** — no local-scope writes or `<gitRoot>/.devflow` resolution anywhere in these changes.

## The 7 gaps

**Gap 1 — Security deny list: full lifecycle** (the careful one)
- New `devflow security` command: `--status` (none/user/managed/both + derived entry count + location), `--enable [--user|--managed]` (add-to-target-first → verify → strip-the-other; self-heals an interrupted switch), `--disable` (strips user + removes managed, **reports exactly which entries were removed**, hard-fails on unparseable settings.json).
- Fresh user-mode init now actually installs the deny list via a dedicated, **atomic**, user-dir step that runs after the managed-install resolves. Re-init injects/updates it (fixes the early-return skip).
- Managed-write **failure now really falls back** to a user-settings write + warn (the old fallback was a hollow message with no write).
- Version drift handled by a frozen `DEVFLOW_HISTORICAL_DENY` superset — detection + removal use the historical set; enable adds the current template. Entry count is always derived, never hardcoded.
- Manifest↔reality conflict: prompt in TTY / keep detected reality + warn in non-TTY / explicit flag always wins.
- `init --no-security` flag; uninstall has a unified security-removal prompt (TTY-confirm default-keep / non-TTY preserve).
- **User-scope-only, machine-global** — always targets `~/.claude/settings.json`, never a project file.

**Gaps 2 & 4 — Decisions/curation runtime gates**
- Shell hooks now gate on **config OR sentinel** (dual-signal), folded once in `dream-evaluate`. `eval-decisions`, `eval-curation`, `dream-capture` (usage scanner), and `dream-collect-tasks` (curation sweep) all honor it. Disabling no longer stamps `.curation-last` (no 7-day suppression after re-enable).

**Gap 3 — HUD hard-disable**
- `hud --disable` removes the Devflow statusLine from settings.json (surgically — a non-Devflow statusLine survives) AND sets `hud.json enabled:false`. Self-healing under drift.

**Gap 5 — Dead memory sentinel removed**
- `getWorkingMemoryDisabledSentinel` removed from all copies + the `.cjs` export; new ENOENT-tolerant per-project migration `purge-dead-working-memory-sentinel-v1`; CLAUDE.md updated to config-only (per ADR-001).

**Gap 6 — Manifest sync + list surfacing**
- `ambient/decisions/hud/memory --enable/--disable` update `manifest.features.*` (only when a manifest already exists — never created as a toggle side-effect). `devflow list` surfaces `security` + `safe-delete` as tri-state (on/off/unknown).

**Gap 7 — safe-delete toggle + queue drain**
- New `devflow safe-delete --enable/--disable/--status` (state lives in the shell-profile block; no-trash-CLI warns + writes nothing; upgrade replaces in place — no duplicate blocks; `$SHELL` unset → "unknown"). `init --no-memory` now drains the pending-turns queue. `.claudeignore` unchanged (user-owned scaffold).

## Cross-cutting
- All `settings.json` writes now go through `writeFileAtomicExclusive` (temp+rename) — no more truncate-on-crash of the file Claude reads every turn.
- New `syncManifestFeature` helper deduplicates toggle→manifest sync across 5 commands.

## Breaking / behavioral changes
- HUD disabled loses the version-update badge (surfaced via `devflow list` instead).
- `eval-decisions` now honors the `.disabled` sentinel (consistency).
- Curation now depends on decisions being enabled.
- `SecurityMode` gains `'none'`.
- No local-scope handling anywhere (clean break).
- Aggressive deny-list removal can strip a user entry that exactly matches a historical Devflow entry — the `--disable`/uninstall removal report lets them re-add.

## Testing
- Full suite green: **1856** tests (excl. the worker integration file) + **42** (worker file, run in isolation) = ~1898 tests, **0 failures**. (The worker-test timeouts seen mid-run were pre-existing machine-load flakiness — those tests are byte-identical to `main` and pass cleanly in isolation.)
- New coverage: `stripUserDenyList` / `detectDenyState` / `resolveSecurityAction` contracts, manifest `security` round-trip, `safe-delete` command, `list` tri-state, shell-hook dual-signal gating, HUD self-healing.
- **Snyk code scan: 0 issues** across `src/cli`.

## Note
- The standalone `security --enable --managed` errors (with guidance) on managed-write failure rather than auto-falling-back — honoring "explicit flag wins". The init path auto-falls-back to user mode. This asymmetry is intentional.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dean0x:
## BLOCKING: Security deny list not written on declined-sudo fallback

**File**: `src/cli/commands/init.ts:1287-1337`
**Confidence**: 95% (Security)

### The Bug
The write step for the managed-mode `securityMode` uses an `if/else if/else if` chain with **no terminal `else`**:
- Line 1287: `if (securityMode === 'managed' && managedSettingsConfirmed)` 
- Line 1315: `else if (securityMode === 'user')`
- Line 1325: `else if (securityMode === 'none')`

When a user selects "Managed settings" but then declines the sudo prompt (line 848), `managedSettingsConfirmed = false` is set while `securityMode` stays `'managed'`. At the write step:
- Line 1287 branch: false (managedSettingsConfirmed is false)
- Line 1315 branch: false (securityMode is still 'managed', not 'user')
- Line 1325 branch: false (securityMode is 'managed', not 'none')
- **No branch executes** → deny list written to neither managed nor user settings

The manifest is then written as `security: 'managed'` (line 1417), falsely recording managed mode while nothing exists on disk.

### Why This Matters
The UI explicitly offers a labeled fallback: "No, fall back to settings.json" (line 839). The user actively selected this, yet receives zero deny-list protection and a manifest/reality mismatch.

### Fix
Route the declined-sudo path to the user-settings write. Either:

**Option A**: Flip the mode at decline time (init.ts:848):
```ts
managedSettingsConfirmed = sudoChoice === 'yes';
if (!managedSettingsConfirmed) {
  securityMode = 'user'; // "fall back to settings.json" must actually write to settings.json
}
```

**Option B**: Add the terminal branch at the write step:
```ts
} else if (securityMode === 'managed' && !managedSettingsConfirmed) {
  // User declined sudo and chose the settings.json fallback — honor it.
  if (templateDeny.length > 0) {
    try {
      await applyUserSecurityDenyList(userSettingsPath, templateDeny);
      securityMode = 'user'; // keep manifest in sync with reality
    } catch (e) {
      p.log.warn(\`Could not write deny list to user settings: ${e instanceof Error ? e.message : e}\`);
    }
  }
}
```

Either approach fixes the reachable gap and keeps the manifest in sync.

dean0x:
## Should-Fix Summary

### Testing Gaps (HIGH confidence — established patterns not followed)

**1. New migration `purge-dead-working-memory-sentinel-v1` has no test** (migrations.ts:1037, 95% confidence)
- Every other file-removing migration in `migrations.test.ts` has a `describe` block with behavioral `.run()` tests
- This migration lacks all three behavioral assertions: success path, ENOENT idempotency, rethrow-on-other-errno
- Fix: Add a `describe('purge-dead-working-memory-sentinel-v1 migration', ...)` block following the `purge-stale-memory-markers-v1` pattern

**2. `applyUserSecurityDenyList` has zero direct tests** (post-install.ts:634, 90% confidence)
- The function has a read-then-merge-then-atomic-write flow with ENOENT-default-to-`{}` branch — completely untested
- This is the **only function that writes the user deny list** and is exercised in two critical init outcomes (normal user mode + managed-write fallback)
- Fix: Add a `describe('applyUserSecurityDenyList')` block with tmpdir tests covering success path, missing-file ENOENT branch, and idempotency

**3. `security.ts` command handler and pure helpers entirely untested** (220-line file, 92% confidence)
- No `tests/security-command.test.ts` exists
- `countDenyEntries` and `loadTemplateDenyEntries` (two pure, easily-testable helpers) lack any coverage
- Fix: Add a `tests/security-command.test.ts` covering the two pure helpers (parse error → 0/[], non-array → 0/[], happy path)

**4. `syncManifestFeature` no-op-when-no-manifest guard is untested** (manifest.ts:113, 85% confidence)
- Documented invariant: "No-op when no manifest exists" — if it regresses to write-anyway, toggles would fabricate partial manifests on un-inited projects
- Fix: Two-line test pair in `manifest.test.ts`: (1) no-op when manifest.json absent, (2) updates field when present

### Structural/Maintainability Issues

**1. TypeScript: `SecurityMode` union lacks `never` exhaustiveness guard** (init.ts:1287/1315/1325, 88% confidence)
- The if/else-if chain compiles fine but if a 4th mode is ever added it silently falls through
- The codebase already establishes the pattern at `migrations.ts:1380` (`const _exhaustive: never = ...`)
- Fix: Convert to a `switch (securityMode)` with a `default: { const _exhaustive: never = securityMode; }` guard

**2. Complexity: `resolveSecurityAction` duplicates mode→target logic 3 times** (post-install.ts:340-410, 88% confidence)
- Lines 380, 386-393, and 407-408 encode the same precedence (both/user → user, managed → managed, none → none/user) but inconsistently
- Fix: Extract one named helper `keepTargetFor(m: DetectedMode)` and call from all three sites, dropping function complexity by ~10 lines

**3. Reliability: `security --disable` throws unguarded on corrupt managed JSON** (security.ts:210-211, 90% confidence)
- The disable path carefully guards user-settings parse (try/catch, graceful exit), but managed-settings removal has no guard
- When managed file is corrupt, `removeManagedSettings` runs and executes `JSON.parse(...)` at post-install.ts:534 with no try/catch → raw crash
- `detectDenyState` already returns `managed=false` for corrupt files, but disable ignores that signal and uses raw file existence instead
- Fix: Gate on the parsed signal or wrap the managed removal to degrade instead of throwing

**4. Performance: `dream-capture` adds jq subprocess before cheap grep gate** (scripts/hooks/dream-capture:91-96, 88% confidence)
- Dual-signal gate calls `json_field_file` (spawns jq/node) on every turn, **before** the cheap citation grep
- Pre-PR the gate was zero subprocesses; this regresses the D29 design intent ("skip Node spawn when no ADR/PF citations present")
- Fix: Reorder — grep first. Move the in-process citation grep ahead of the config parse so expensive parse only runs when a citation is actually present

**5. Consistency: `security` and `safe-delete` diverge from toggle-command conventions** (security.ts/safe-delete.ts, 90% confidence)
- New commands treat "no flag" as show-status; all four pre-existing toggles treat it as show-Usage and return
- New commands hard-exit on operational failures; siblings degrade with warn + return
- Fix: Pick one convention (safest: make new commands match incumbents on no-flag behavior)

**6. Architecture: `SecurityMode` union declared in 7 places instead of 1** (post-install.ts:624 + 6 others, 90% confidence)
- Exported canonical type exists but inlined in resolveSecurityAction signature, manifest.ts, and InitOptions
- Adding a 4th mode would require touching ~7 sites with no compiler enforcement of consistency
- Fix: Use the single exported `SecurityMode` type everywhere (import + reuse)

**7. Documentation (HIGH): CLI reference missing two new commands and init flag** (docs/cli-reference.md, 96% confidence)
- No `## Security` or `## Safe-Delete` sections in the canonical reference
- Init Options table missing `--security <user|managed|none>` flag
- Fix: Add two sections with usage examples + one table row for the init flag

---

### Note on PR Description Claim
The PR states "all settings.json writes are now atomic," but this is **broader than reality**:
- ✅ All security-path writes use `writeFileAtomicExclusive`
- ❌ Five pre-existing plain `fs.writeFile` remain on HUD/viewMode init+uninstall paths (lines 251, 1173, 279, 382, 399 — not introduced here)

This PR correctly establishes the invariant for the security seam; a follow-up PR should extend it repo-wide.

dean0x:
## Additional Should-Fix & Lower-Confidence Findings

### Additional Type-Safety Issues (MEDIUM, 82-88% confidence)

**1. `resolveSecurityAction` recomputes DetectedMode from same booleans** (post-install.ts:358-362, 82% confidence)
- The 4-way ternary deriving `detectedMode` embeds boolean logic mid-function that callers already compute
- Fix: Lift the mapping to a tiny named function `classifyDetected(detected): DetectedMode` and reuse it

**2. `TriState` unions lack `never` exhaustiveness** (list.ts:38-42, safe-delete.ts:74-80, 82% confidence)
- `triLabel` and the `SafeDeleteStatus` chain use trailing-`else` defaults instead of checked exhaustiveness
- If `TriState` gains a 4th state they'll silently render as `unknown` with no compiler signal
- Fix: Use switch with `default: { const _exhaustive: never = state; ... }` or add `const _: never = state` assertion

### Reliability & Fault Tolerance (MEDIUM, 85-90% confidence)

**1. Strip-path duplication: 4 near-identical read-strip-write blocks** (init.ts + security.ts, 80% confidence)
- The sequence `read user settings → stripUserDenyList → guard changed → writeFileAtomicExclusive → swallow ENOENT` appears verbatim in 4 places
- The merge side got `applyUserSecurityDenyList` helper; the strip side wasn't given the same treatment
- Fix: Extract `stripUserSecurityDenyList(settingsPath)` helper alongside `applyUserSecurityDenyList` and reuse everywhere

**2. Template deny-list loading duplicated in 4 places** (security.ts + init.ts + post-install.ts ×2, 88% confidence)
- `loadTemplateDenyEntries` helper exists in security.ts but isn't reused by the 3 other inline copies
- Fix: Move `loadTemplateDenyEntries` to post-install.ts alongside other deny-list helpers and import everywhere

**3. `stripUserDenyList` / `mergeDenyList` throw on corrupt input (latent)** (post-install.ts, 85% confidence)
- Both pure helpers call `JSON.parse(...)` with no try/catch
- Documented as "PURE + idempotent" but not as partial functions that throw
- All current callers pre-validate or wrap in try/catch (safe), but future callers will trust the comment and skip the guard
- Fix: Either document the throw explicitly in JSDoc or return a Result/null-on-parse-failure shape per the parse-at-boundaries rule

### Pre-Existing Issues Surfaced by New Code

**1. HUD `--enable` not hardened against corrupt settings.json** (hud.ts, pre-existing but adjacent to hardened disable path, 88% confidence)
- The NEW `hud --disable` self-heal correctly wraps its call in try/catch
- But `hud --enable` calls `hasHudStatusLine(...)` outside any try/catch — malformed settings makes it crash with raw stack
- Inconsistent with this PR's theme of never crashing on settings.json
- Fix (optional, mirrors disable hardening): wrap the settings parse in try/catch in the enable path

**2. Uninstall security comment misleads about hard-fail behavior** (uninstall.ts:461 comment vs :464-468 code, 90% confidence)
- Comment claims "hard fail if unparseable — mirrors security --disable"
- But code does NOT hard-fail — `detectDenyState` sets `user=false` for unparseable input, so strip is skipped (no-op, no error)
- This is **opposite** of `security --disable` (which calls `process.exit(1)`)
- Fix: Either correct the comment to accurately describe the no-op-on-corrupt behavior, or add explicit hard-fail to match the comment

### Documentation Gaps (MEDIUM-HIGH, 85-96% confidence)

**1. README omits new public commands** (README.md:111-123, 88% confidence)
- Lists existing toggles like `rules --status` and `decisions --enable` but not `security` or `safe-delete`
- Fix: Add two lines to the `## CLI Reference` block

**2. CLAUDE.md migrations enumeration incomplete** (CLAUDE.md:68, 85% confidence)
- Lists 4 existing migrations but omits the new `purge-dead-working-memory-sentinel-v1`
- Fix: Append to the migrations list

### Lower-Confidence Observations (60-79% range)

- **Dynamic import of already-static module** (init.ts:895, 78%) — imports `getManagedSettingsPath` from `../utils/paths.js` which is already statically imported; unnecessary async hop
- **`SecurityMode` re-declared in 7 places** (90% reported separately above)
- **Template-deny loading not hoisted** (88% reported separately above)
- **Dual-signal gate split across two logic branches** (session-start-context:47 vs :122-132, 70%) — acceptable in practice (both paths set the signals), but latent inconsistency if a future change sets config-only

---

## Summary of Issues by Category

| Type | Count | Severity |
|------|-------|----------|
| **BLOCKING** | 1 | Security deny list not written on declined-sudo |
| **Should-Fix (Testing)** | 4 | Missing test coverage for new paths |
| **Should-Fix (Type-Safety)** | 3 | Missing `never` exhaustiveness guards + union redeclaration |
| **Should-Fix (Maintainability)** | 7 | Duplication, structural issues, consistency divergences |
| **Should-Fix (Reliability)** | 3 | Unguarded throws, insufficient fault tolerance |
| **Should-Fix (Documentation)** | 2 | CLI reference and CLAUDE.md gaps |
| **Pre-Existing** | 2 | HUD enable, uninstall comment (adjacent to new code) |
| **Low-Priority** | 3 | Style and minor structural nits |

**No findings block merge once the BLOCKING security issue is fixed. Most should-fixes are low-effort consolidations (extract helpers, add guards, fix comments).**

---

Reviewed with: security.md, reliability.md, performance.md, complexity.md, consistency.md, testing.md, typescript.md, documentation.md, architecture.md, regression.md

PR #256, opened 2026-07-10
feat(decisions): move decisions index from runtime script to write-time artifact

dean0x:
## Summary

- **Removes** `decisions-index.cjs` subprocess — every command surface that previously shelled out to `node ~/.devflow/scripts/hooks/lib/decisions-index.cjs index "<worktree>"` now does a plain `Read` of `.devflow/decisions/index.md`
- **`render-decisions.cjs`** writes `index.md` alongside `decisions.md`/`pitfalls.md` at every render (Phase 1)
- **14 command surfaces** updated: 9 static MDS hosts (`implement`, `plan`, `code-review`, `resolve`, `debug`, `self-review`, `explore`, `research`, `bug-analysis`) + 3 dynamic MDS hosts + `_preamble.mds` partial + `release.md` (Phase 2)
- **Bootstrap migrations** added: `render-decisions-index-v1` (per-project, generates `index.md` from existing ledger on `devflow init`) and `purge-orphaned-decisions-index-v1` (global, removes the installed `decisions-index.cjs` copy from `~/.devflow/scripts/hooks/lib/`) (Phase 3)
- **`decisions-index.cjs` deleted** from source; tests retargeted to `buildIndexContent`/`selectActiveRows` in-process pipeline; docs updated (Phase 4)

## Changes

- `scripts/hooks/lib/render-decisions.cjs` — extends `renderAndWriteAll` to write `index.md` last; extends `--check` mode to include `index.md` drift; adds `selectActiveRows` export; adds `safePath()` + null byte validation at CLI boundary
- `scripts/hooks/lib/decisions-format.cjs` — adds `buildIndexContent(activeDecisionRows, activePitfallRows, opts)` (byte-compat with old `loadDecisionsIndex()` output, validated by TRANSITIONAL GOLDEN test)
- `scripts/hooks/lib/project-paths.cjs` + `src/cli/utils/project-paths.ts` — adds `getDecisionsIndexPath`
- `commands/_partials/_decisions.mds` — new partial encapsulating the index.md read step
- `src/cli/utils/decisions-ledger-migration.ts` — adds `renderDecisionsIndex()` for bootstrap migration
- `src/cli/utils/migrations.ts` — registers two new migrations
- `tests/decisions/index-content.test.ts` — replaces deleted `index-generator.test.ts`; tests `selectActiveRows + buildIndexContent` pipeline with belt-and-suspenders active-only filter coverage and `renderAndWriteAll` integration
[...]

PR #257, opened 2026-07-11
fix(reliability): bound acquireMkdirLock loop when stale lock is un-removable

dean0x:
## Summary
`acquireMkdirLock` (`src/cli/utils/mkdir-lock.ts`) could spin at 100% CPU indefinitely, escaping its own `timeoutMs`. The stale-break path did `rmdir` then a bare `continue`, which jumped back to the top of the `while (true)` loop and skipped BOTH the timeout guard and the 100ms backoff. If a lock dir was simultaneously **stale** (old mtime) and **un-removable** (`rmdir` fails — e.g. a non-empty dir → `ENOTEMPTY`, or a permissions error), the loop never terminated. This violated the "every loop/retry has a fixed upper bound" reliability rule.

## Fix
Track whether `rmdir` succeeded. Only `continue` (retry immediately) on a successful stale-break; on failure, fall through to the existing timeout guard + 100ms backoff so the loop stays bounded and returns `false` after `timeoutMs` (the correct fail-closed outcome — callers like `assign-anchor` then don't proceed). No change to the signature, the happy path, or the normal-contention path.

## Test
Added a regression test that makes the lock dir un-removable by placing a file inside it (`rmdir` → `ENOTEMPTY`) with `staleMs=1`. Against the old code the test **times out at 3000ms** (proving the spin); with the fix it returns `false` at ~305ms. Full suite: 1924 tests pass, 0 TypeScript warnings; shared-primitive consumers (queue-append, memory refresh, decisions ledger-ops) all green — no regressions.

## Context
The lock mechanism itself is warranted and unchanged — it serializes concurrent `assign-anchor` numbering and multi-file ledger rendering across processes and languages (.cjs helpers + .ts CLI). This fixes only the unbounded-loop edge case. Split out of the PR #256 code review (surfaced there as ISSUE-23, deferred from that PR to keep it focused).

### DIFF
--- a/src/cli/utils/mkdir-lock.ts
+++ b/src/cli/utils/mkdir-lock.ts
@@ -32,8 +32,13 @@ export async function acquireMkdirLock(lockDir: string, timeoutMs = 30_000, stal
       try {
         const stat = await fs.stat(lockDir);
         if (Date.now() - stat.mtimeMs > staleMs) {
-          try { await fs.rmdir(lockDir); } catch { /* race condition OK */ }
-          continue;
+          // D57: Track whether rmdir succeeded. Only retry immediately on success.
+          // If the lock dir is un-removable (e.g. ENOTEMPTY, permission error) we must
+          // fall through to the timeout guard + backoff below — never spin unconditionally.
+          let removed = true;
+          try { await fs.rmdir(lockDir); } catch { removed = false; /* race condition OK */ }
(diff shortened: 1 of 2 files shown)


dean0x:
Superseded — the mkdir-lock stale-spin fix was folded into `refactor/move-decisions-index` (PR #256) as commit 451285a and ships with that PR instead of landing standalone. Closing this and deleting the branch.

PR #260, opened 2026-07-12
feat: productionalize skill/rule shadowing

dean0x:
## Summary

File-copy has been the only installer for two releases. This PR removes the dead native CLI code path, drops the now-pointless SHADOW_RENAMES migration machinery, hardens shadow validation with user-visible install reporting, adds full `rules shadow/unshadow/list` CLI parity, adds an uninstall rule-shadow warning, and documents all of the above.

> **BREAKING CHANGE**: `devflow init` no longer registers the Claude plugin marketplace entry (`extraKnownMarketplaces`) or installs via the native `/plugin install` CLI path. The file-copy installer is the sole installation mechanism.
>
> **Pre-existing condition**: Users who previously ran `devflow init` with the native CLI path still have marketplace-registered plugins. The file-copy uninstaller (`devflow uninstall`) never touched those marketplace entries and continues not to — users wishing to fully remove marketplace registrations must do so manually in Claude Code settings.

### Step 1 — Remove native CLI install path (`6e1efb9`)
- Deleted `src/cli/utils/cli.ts` (sole export was `isClaudeCliAvailable`)
- Deleted `addMarketplaceViaCli`, `installPluginViaCli`, `installViaCli` from `installer.ts`
- Removed `uninstallPluginViaCli` and CLI-attempt block from `uninstall.ts`
- Removed `extraKnownMarketplaces` block from `src/templates/settings.json` and `docs/reference/file-organization.md`

### Step 2 — Drop SHADOW_RENAMES and shadow-overrides migration (`b493f64`)
- Deleted `src/cli/utils/shadow-overrides-migration.ts`
- Deleted `SHADOW_RENAMES` export from `plugins.ts`
- Removed `shadow-overrides-v2-names` migration from `migrations.ts`
- Deleted `tests/shadow-overrides-migration.test.ts`; updated test allowlists accordingly

### Step 3 — Harden shadow validation with `InstallReport` API (`e4f77e1`)
- `validateSkillShadow` / `validateRuleShadow` — new exported helpers; the `isFile()` guard in `validateRuleShadow` prevents EISDIR crashes in the `Promise.all` rules block
- `installViaFileCopy` now returns `InstallReport` (shadowed skills, shadowed rules, skipped shadows with typed reasons)
- `installRuleFile` returns `RuleInstallOutcome` discriminated string type
- `devflow init` summary reports applied shadows (info) and invalid-shadow skips (warn)

### Step 4 — `rules shadow/unshadow/list` CLI; uninstall rule-shadow warning (`d4e2313`)
- `rules.ts`: positional `shadow <name>` (seeds from installed rule, falls back to plugin source), `unshadow <name>`, `list` (with shadow/invalid-shadow annotations); exported `hasRuleShadow` and `listShadowedRules`
- `skills.ts`: renamed `list-shadowed` → `list`; annotates each skill as shadowed/shadowed-invalid/unshadowed
- `uninstall.ts`: warns on both skill and rule shadows before uninstalling

### Step 5 — Documentation (`1e7eb69`)
- `docs/cli-reference.md`: new `rules shadow/unshadow/list` entries; Rule Shadowing section; `skills list` rename
- `CLAUDE.md` + `README.md`: mention rule shadow CLI
- 6 plugin READMEs: remove "Via Claude Code (when available)" `/plugin install` section (ADR-003: end-state only)

## Breaking Changes

`devflow init` no longer calls `/plugin install` or registers `extraKnownMarketplaces`. Users relying on the native Claude CLI install path for fresh installs must switch to `npx devflow-kit init`. Existing marketplace-registered plugins from prior installs are unaffected.

## Reviewer Focus Areas

- `src/cli/utils/installer.ts` — `validateRuleShadow` EISDIR guard and `InstallReport` accumulation logic
- `src/cli/commands/rules.ts` — shadow seed two-stage lookup (installed → plugin source fallback)
- `src/cli/commands/skills.ts` — `list` action validity annotation loop
- `tests/rules.test.ts` / `tests/skill-namespace.test.ts` — new shadow-validation and report coverage


dean0x:
## Issue: Lossy `installRuleFile` return forces redundant re-stat

**File**: `src/cli/utils/installer.ts`  
**Lines**: 353-358  
**Confidence**: 88% (flagged by 4 reviewers)  
**Severity**: MEDIUM (BLOCKING)

The `installRuleFile` function calls `validateRuleShadow(shadowFile)` and already knows the exact reason (`'empty-shadow-file'` vs `'not-a-file'`), but collapses both into a single `RuleInstallOutcome` value `'source-invalid-shadow'` (line 120). The caller must re-stat the file to recover what was just known:

```typescript
} else if (outcome === 'source-invalid-shadow') {
  const shadowFile = path.join(devflowDir, 'rules', \`${ruleName}.md\`);
  const shadowState = await validateRuleShadow(shadowFile);   // redundant re-stat
  const reason = shadowState === 'not-a-file' ? 'not-a-file' : 'empty-shadow-file';
  report.skippedShadows.push({ kind: 'rule', name: ruleName, reason });
}
```

**Fix**: Return the reason from `installRuleFile` instead of collapsing it:
```typescript
export type RuleInstallOutcome =
  | { kind: 'shadow' }
  | { kind: 'source' }
  | { kind: 'source-invalid-shadow'; reason: 'empty-shadow-file' | 'not-a-file' }
  | { kind: 'skipped' };
```

dean0x:
## Issue: Unused local `shadowDirSet` (dead code)

**File**: `src/cli/commands/skills.ts`  
**Line**: 124  
**Confidence**: 95%  
**Severity**: MEDIUM

`const shadowDirSet = new Set(shadowDirNames);` is declared but never read. The code uses `knownSkillSet` and the raw `shadowDirNames` array instead.

**Fix**: Delete line 124. (This will also be fixed by the refactor in Comment 4 if those changes are applied together.)

dean0x:
## Issue: `rulesCommand.action` exceeds complexity limits

**File**: `src/cli/commands/rules.ts`  
**Lines**: 108-269  
**Confidence**: 85%  
**Severity**: HIGH (BLOCKING)

The single `.action` callback is a 161-line multi-mode dispatcher handling both positional actions (`shadow`/`unshadow`/`list`/unknown) and flag actions (`--enable`/`--disable`/`--status`/`--list`) in one body. Nesting depth reaches 5 blocks deep, and decision-point count exceeds 20. This fails the project thresholds (function >50 lines + nesting >4 = HIGH) and the explainability test.

**Fix**: Extract per-action handlers, starting with the two-stage seed logic for the `shadow` branch:
```typescript
async function seedRuleShadow(name: string, rulesTarget: string, devflowDir: string): Promise<boolean> {
  // Seed from installed, then fall back to plugin source
}
async function handleRuleShadow(...) { ... }
async function handleRuleUnshadow(...) { ... }
async function handleRuleEnable(...) { ... }
```
Then the `.action` callback becomes a thin router delegating to these handlers.

dean0x:
## Issue: `skills list` re-stats every skill sequentially, ignoring readdir cache

**File**: `src/cli/commands/skills.ts`  
**Lines**: 130-143  
**Confidence**: 90%  
**Severity**: MEDIUM (BLOCKING)

Two compounding inefficiencies in the skill list loop:

1. **Sequential when parallel possible**: The `for (const skill of allSkills)` loop awaits `validateSkillShadow` one skill at a time, serializing ~80 stat round-trips. The sibling `printRulesList` in rules.ts uses `Promise.all`.

2. **Redundant I/O**: `shadowDirSet` (line 124, see Comment 2) already lists every shadow directory. But the loop calls `validateSkillShadow` for every skill, including ~38 unshadowed ones. Each does an `fs.stat` that resolves to ENOENT.

**Fix**: Short-circuit on `shadowDirSet` and parallelize:
```typescript
const rows = await Promise.all(
  allSkills.map(async (skill) => {
    let statusTag = color.dim('—');
    if (shadowDirSet.has(skill)) {
      const shadowState = await validateSkillShadow(path.join(shadowsRoot, skill));
      statusTag = shadowState === 'valid' ? color.green('shadowed') : color.yellow('shadowed — invalid: ...');
    }
    return \`  ${color.cyan(skill.padEnd(28))} ${statusTag}\`;
  }),
);
```

dean0x:
## Issue: CHANGELOG `[Unreleased]` section missing breaking-change and feature entries

**File**: `CHANGELOG.md`  
**Line**: 8  
**Confidence**: 88% (regression) / 80% (documentation)  
**Severity**: MEDIUM (BLOCKING per regression review)

The PR claims its breaking changes are documented, but the `[Unreleased]` section omits them:

- **BREAKING**: `skills list-shadowed` renamed to `skills list` (users scripting the old name now hit "Unknown action")
- **Added**: `rules shadow` / `rules unshadow` / `rules list` commands
- **Removed**: Native `claude plugin install` path, `extraKnownMarketplaces` marketplace entry, `SHADOW_RENAMES` migration machinery

**Fix**: Add entries to the `### Changed` / `### Added` / `### Removed` sections:
```markdown
### Changed
- `skills list-shadowed` → `skills list` (BREAKING)

### Added
- `rules shadow <name>` / `rules unshadow <name>` / `rules list` commands

### Removed
- Native plugin install path (`claude plugin install`)
- `extraKnownMarketplaces` marketplace registry
- `SHADOW_RENAMES` migration machinery
```

dean0x:
## Issue: `rules unshadow` skips allowlist validation that `rules shadow` enforces

**File**: `src/cli/commands/rules.ts`  
**Lines**: 167-179  
**Confidence**: 90%  
**Severity**: LOW (Should Fix)

The `shadow` action validates `allRules.includes(name)` (line 124) before touching the filesystem. The `unshadow` action does not — it goes straight to `fs.rm(shadowFile, { force: true })` (line 179).

While this is not a trust-boundary crossing (the operator's own input, same privileges), it's a defense-in-depth / consistency gap: both branches should validate identically (see PF-003 in decisions).

**Fix**: Mirror the guard:
```typescript
} else if (action === 'unshadow') {
  if (!name) { /* ...existing error... */ }
  if (!allRules.includes(name)) {
    p.log.error(\`Unknown rule: ${name}\`);
    p.log.info(\`Available rules: ${allRules.join(', ')}\`);
    process.exit(1);
  }
  const shadowFile = path.join(devflowDir, 'rules', \`${name}.md\`);
  // ...unchanged...
}
```

dean0x:
## Issue: Non-exhaustive `ShadowSkipReason` handling allows silent mislabeling

**File**: `src/cli/commands/init.ts`  
**Line**: 1365  
**Confidence**: 90%  
**Severity**: MEDIUM (BLOCKING)

The skip-message ternary handles only 2 of 3 `ShadowSkipReason` members explicitly:
```typescript
const reasonMsg = skip.reason === 'missing-skill-md' ? '...'
  : skip.reason === 'empty-shadow-file' ? '...'
  : 'shadow path is not a file';   // catches 'not-a-file' AND future additions
```

If a fourth reason is added to the union, this branch silently mislabels it as "shadow path is not a file" with no compile error.

**Fix**: Use an exhaustive switch with a `const _: never = reason` guard:
```typescript
function reasonMessage(reason: ShadowSkipReason): string {
  switch (reason) {
    case 'missing-skill-md': return 'shadow directory has no valid SKILL.md';
    case 'empty-shadow-file': return 'shadow file is empty';
    case 'not-a-file': return 'shadow path is not a file';
    default: { const _: never = reason; throw new Error(\`Unhandled reason: ${_}\`); }
  }
}
```

dean0x:
## Issue: Malformed comment alignment in bash code block

**File**: `docs/cli-reference.md`  
**Line**: 109  
**Confidence**: 95%  
**Severity**: LOW

The line `npx devflow-kit rules unshadow <name># Remove rule override` has no space between `<name>` and the `#` comment. Every sibling line in the block aligns its `#` comment with padding. Without preceding whitespace, `#` is not a shell comment token.

**Fix**: Add spacing to match neighbors:
```
npx devflow-kit rules unshadow <name> # Remove rule override
```

Issue #255, opened 2026-07-09
HUD/CI: test files (tests/**) are transpile-only and outside tsconfig `include` — TS regressions in tests escape CI


dean0x:
## Problem

\`tsconfig.json\` scopes \`include\` to \`src/cli/**\`, and vitest transpiles tests via esbuild without type-checking. As a result, TypeScript regressions inside \`tests/**\` are not caught by \`npm run build\` or CI.

## Evidence

Surfaced during code review of PR #254 (chore/devflow-fine-tuning):
- **Reviewer**: typescript reviewer
- **Confidence**: 90%
- **Classification**: pre-existing / out-of-blast-radius (Triager verdict)

## Scope

This is a pre-existing, project-wide CI/tsconfig concern unrelated to PR #254's four-tweak purpose. Fixing it would require a project-wide tsconfig or CI change, so it was deferred to its own ticket rather than expanded into that PR.

## Suggested Approaches

1. Add a \`typecheck:tests\` script: \`tsc --noEmit -p tsconfig.test.json\` covering \`tests/**\`, wired into CI
2. Widen the existing tsconfig.json \`include\` to cover both \`src/cli/**\` and \`tests/**\`

## Related

Relates to PR #254 (chore/devflow-fine-tuning)

PR #269, opened 2026-07-24
feat: external model routing (devflow proxy) + per-agent model config (devflow agents)

dean0x:
## Summary

Adds integrated support for routing Devflow agents through GPT models via an OpenAI/Codex subscription:

- **`devflow proxy`** (`--enable/--disable/--status`): routes GPT model IDs through a local relay by injecting `ANTHROPIC_BASE_URL` into `settings.json`; all other models pass through to Anthropic. Default OFF, Advanced-init only.
- **`devflow agents`** (interactive TUI + `--list/--set/--reset`): per-agent model and effort overrides, stored deviations-only in `~/.devflow/agent-models.json` and converged into installed agent frontmatter.

The routable model set is **discovered live** from the relay binary and cached under `~/.devflow/cache/models/` (24h TTL, versioned keys, stale entries serve as fallback when discovery fails). There is no hardcoded model registry.

## Key Design Points

- **Preflight is static-only**: enable validates four static prerequisites — relay binary resolvable, Codex auth present, port free or already a Devflow relay, `settings.json` parseable with no foreign `ANTHROPIC_BASE_URL`. Live verification (the relay's own `doctor`) runs *after* spawn, because a check that requires the relay to be up cannot gate its creation.
- **Rollback tears down only what this invocation created**: a self-spawned relay is killed on verification failure; an adopted relay already serving other sessions never is.
- **Dormancy**: GPT mappings are stored while routing is off and materialize into agent frontmatter only when routing is enabled. Effort overrides are orthogonal and always apply.
- **Disable preserves the mapping**: reverts agent frontmatter to Claude defaults but remembers assignments so re-enable restores prior state.
- **Relay lifecycle**: never killed by disable or uninstall (loopback-only, dies at reboot); the `ensure-proxy` hook auto-revives it at session start within a bounded window.
- **Init resilience**: init runs the same preflight but never spawns or runs doctor; preflight failure warns and forces the feature off rather than aborting init.
- **Effect timing**: changes take effect in new Claude Code sessions, not mid-session.

## Implementation

**New core modules** (agent-neutral, per ADR-013):
- `src/core/external-models.ts` — Claude alias table, dormancy predicate
- `src/core/model-discovery.ts` — live catalog discovery, cache selection, stale fallback
- `src/core/cache.ts` — cache envelope read/write, `0700`/`0600` permissions, TTL clamping
- `src/core/proxy-log.ts` — relay env allowlisting, log handling
- `src/core/proxy-state.ts` — runtime state I/O
- `src/core/codex-auth-inspect.ts` — pure auth inspector (`absent | unreadable | present`)
- `src/core/agent-frontmatter.ts` / `src/core/agent-models.ts` — frontmatter rewrite engine, schema, apply/revert

**CLI and hooks**:
- `src/cli/commands/{proxy,agents}.ts` — command entrypoints
- `src/cli/agents-view/` — pure reducer + renderer + thin terminal shell for the TUI
- `src/assets/scripts/hooks/ensure-proxy` — bash 3.2-safe, bounded, exit-0 relay revival hook (SessionStart + UserPromptSubmit)
- Init and uninstall wiring; docs across README, `docs/cli-reference.md`, `CLAUDE.md`, and the feature knowledge base

**Dependency**: `subswitch@0.2.0`, exact-pinned.

## Review History

Two full review cycles ran on this branch.

**Cycle 2** (12 reviewers, 55 files): 77 issues triaged — 69 fixed, 4 by design, 4 deferred to follow-up issues (#278–#281), 0 false positives. Three themes: partial security hardening, vacuous tests, and documentation drift. Notable fixes include converting the relay-spawn env handling from a denylist to an `env -i` allowlist across all four spawn sites, de-vacuuming a batch of tests that passed unconditionally, and reconciling documentation with the deleted model registry.

**Post-review manual verification** surfaced two further defects, both fixed here:
- The interactive TUI suppressed all external models when routing was disabled, via two independent gates (`agents.ts` never consulted a catalog on the proxy-off path; `buildModelCycle` short-circuited on `!proxyEnabled`). This made the dormant-mapping flow unreachable through the primary UI — a user could not create the very state the TUI was built to display. Fixed by sourcing the cached catalog on the proxy-off path (synchronous, zero spawns) and removing the redundant gate. Acceptance criterion AC-F5, which had ratified the broken behavior and predated the model-discovery work, was revised.
- The relay-spawn allowlists omitted `NODE_EXTRA_CA_CERTS`, breaking users behind a TLS-intercepting corporate proxy. Added to both the shell and TypeScript allowlists, with a drift guard asserting the two stay in sync. `NODE_OPTIONS` remains deliberately excluded — it permits arbitrary code execution via `--require`.

## Quality Gates

- **Build**: `npm run build` clean — 14 MDS commands compiled, 0 errors, 0 warnings
- **Typecheck**: `tsc --noEmit` clean
- **Tests**: 84 files, 2,673 tests, 0 failures
- **Snyk**: `security/snyk` CI check passing
- **Pipeline**: Validator → Simplifier → Scrutinizer (9-pillar) → Evaluator → QA Tester (32 black-box scenarios)

## Test Plan

- **Unit**: model discovery and cache selection, proxy state transitions, agent model resolution, frontmatter rewriting, TUI reducer and renderer
- **Shell hook**: `ensure-proxy` behavior, relay revival, env allowlisting in the spawned process
- **Integration**: enable/disable/re-enable, frontmatter convergence, init and uninstall wiring, preflight-failure convergence
- **Black-box E2E**: 32 scenarios across enable, disable, session restart, relay crash recovery, port conflicts, auth failures, env cleanup
- **Manual (maintainer, real TTY)**: `--enable`, `--disable`, `--status`, `agents --list`, and the interactive TUI

## Known Follow-ups

Filed, not blocking: #278 (extract proxy status formatters), #279 (split the `agents.ts` handler), #280 (`parseSettings` helper), #281 (`--list` availability display — see the comment there about the existing test's typing gap, which interacts with #255).


dean0x:
## Findings Summary

This PR adds comprehensive external model routing support. The code quality is generally strong with well-defended security boundaries and pure/testable architecture. The review found several issues worth addressing before merge:

### HIGH-Confidence Blocking Issues (≥80%)

**1. Remembered-port logic is dead** (proxy.ts:492, TypeScript Review, 85% confidence)

The `--port` option declares a default value, which defeats the remembered-port feature:

```typescript
.option('--port <n>', 'Port for the local relay (default: 4141)', String(DEFAULT_PROXY_PORT))
```

Because Commander fills `options.port = '4141'` when the flag is absent, the logic at lines 682-687 that checks `if (portOption !== undefined)` will always execute. The prior port from `proxy.json` is computed at line 673 but immediately discarded.

**Expected behavior**: `devflow proxy --enable --port 5000`, disable, then re-enable without `--port` should re-enable on port 5000.
**Actual behavior**: Enables on 4141 instead, spawning a second relay.

**Fix**: Drop the default so omission is detectable:
```typescript
.option('--port <n>', 'Port for the local relay (default: remembered or 4141)')
// portOption stays undefined → priorPort is used
```
---

**2. Preflight production implementations duplicated** (Architecture Review, 90%; Complexity Review, 88%)

The three production preflight dependencies (`realTcpConnectable`, `realHttpGet`, `realSpawnDoctor`) are implemented identically in two places:
- `src/cli/commands/proxy.ts` lines 402–476 (module-private functions)
- `src/cli/commands/init.ts` lines 1258–1301 (inline inside preflight block)

The comment at init.ts:1257 acknowledges this: "see proxy.ts for identical patterns". This creates:
- **DRY violation**: A timeout or error-handling fix applied in one will silently drift from the other
- **Bloated init.ts**: Already a large orchestrator file, now with duplicated I/O code
- **Import duplication**: init.ts pulls in `net`/`http`/`https`/`spawn` purely to reconstruct behavior already in proxy.ts

**Fix**: Export a factory from proxy.ts:
```typescript
export function buildRealPreflightDeps(
  settingsPath: string,
  onWarn?: (msg: string) => void,
): ProxyPreflightDeps {
  return {
    resolveProxyBin,
    fileExists: async (p) => { try { await fs.access(p); return true; } catch { return false; } },
    tcpConnectable: realTcpConnectable,
    httpGet: realHttpGet,
    readSettingsJson: () => fs.readFile(settingsPath, 'utf-8').catch(() => '{}'),
    spawnDoctor: realSpawnDoctor,
    onWarn,
  };
}
```

Then both `runEnable` (proxy.ts) and init's preflight block call it.

---

**3. Inconsistent CLI error exit codes** (Consistency Review, 85%)

The `devflow proxy` command exits 0 on errors (lines 683–686 and others), while the sibling `devflow agents` command (same PR) always sets `process.exitCode = 1` on failures:

```typescript
// proxy.ts - exits 0 even on preflight failure
if (isNaN(parsed) || parsed < 1 || parsed > 65535) {
  p.log.error(`Invalid port: ${portOption}`);
  return;  // ← no exit code
}
```

This breaks failure detection in CI or scripts. Impact: `devflow proxy --enable` in a CI pipeline that fails preflight still exits 0.

**Fix**: Set `process.exitCode = 1` before `return` on all hard-failure paths, matching `agents.ts`.

---

**4. ensure-proxy hook sources json-parse without diagnostic** (Consistency Review, 82%)

Line 28 of ensure-proxy:
```bash
source "$SCRIPT_DIR/json-parse" || { exit 0; }
```

All 11 sibling hooks (including `preamble`, the closest UserPromptSubmit sibling) use:
```bash
source "$SCRIPT_DIR/json-parse" || { echo "<hook>: failed to source json-parse" >&2; exit 1; }
```

The ensure-proxy form is silent, making a broken hook invisible during debugging. The `exit 0` fail-open is defensible given proxy's optional nature, but the missing diagnostic message is a pure inconsistency.

**Fix**: Emit the diagnostic:
```bash
source "$SCRIPT_DIR/json-parse" || { echo "ensure-proxy: failed to source json-parse" >&2; exit 0; }
```

---

**5. ensure-proxy uses curl despite claiming portability** (Consistency Review, 80%)

Line 110–111 (TCP probe) avoids curl:
```bash
# TCP probe (bash built-in, no nc/curl, Bash 3.2-safe)
exec 3<>/dev/tcp/127.0.0.1/$_p 2>/dev/null
```

But line 131 unconditionally uses curl:
```bash
HEALTH_BODY=$(curl -s --max-time 2 "http://127.0.0.1:${PROXY_PORT}/__subswitch/health")
```

This creates:
- **Portability deviation**: Within one file the author avoids curl for portability, then depends on it
- **Unguarded dependency**: No `command -v curl` check; on a system without curl, substitution yields empty body, triggering a false "port squatted" warning even when the relay is ours

**Fix**: Either drop the health-check identity check on the hook side (the CLI `--status` already does a real check) or use /dev/tcp for the GET, matching the file's stated philosophy. At minimum:
```bash
if command -v curl >/dev/null 2>&1; then
  HEALTH_BODY=$(curl -s --max-time 2 "http://127.0.0.1:${PROXY_PORT}/__subswitch/health")
  # ... health check
else
  # Assume ours if curl unavailable
  :
fi
```

---

### Additional Findings

**Medium-Confidence Items (60–79% confidence)** are summarized in the full review reports and available for reference. High-level:
- Long functions (`runEnable` 182 lines, `runStatus` 142 lines) are extraction opportunities
- Scattered magic timeouts (2000, 10_000, 50×100ms) merit named constants
- Dormancy predicate (`isGpt && !proxyEnabled`) re-expressed at 4 sites
- Spawn error listeners missing (affects crash-safety under fd exhaustion)
- Test coverage gaps in terminal.ts, proxy-state.ts, and runEnable relay-spawn paths

All review reports are available in `.devflow/docs/reviews/feat-external-model-routing/2026-07-24_2302/`.

---

**Next Steps**: Address the 5 blocking items above, then re-run the review to verify fixes. The codebase quality is strong; these are refinements to robustness and consistency.

CC devflow-kit maintainer | Review cycle 1

dean0x (commented):
**Line 492: Remembered-port logic is dead** (TypeScript Review, 85%)

The default value defeats the remembered-port feature. When `--port` is omitted, Commander fills `options.port = '4141'`, making the check at line 680 (`if (portOption !== undefined)`) always true. The prior port from `proxy.json` (computed at line 673) is discarded.

Expected: Re-enable on the remembered port
Actual: Re-enables on 4141, spawning a second relay

**Fix**: Drop the default:
```
.option('--port <n>', 'Port for the local relay (default: remembered or 4141)')
```

dean0x:
## Code Review — Inline Comments Summary

I've reviewed the 12 reviewer reports and created inline comments on the key findings (≥80% confidence). Here's what was identified and commented:

### Critical/High-Severity Findings (Blocking)

**1. `src/assets/scripts/hooks/ensure-proxy:246`** — Relay spawn missing allowlist (95% confidence, HIGH)
- Primary production spawn path still uses denylist instead of the allowlist applied to 3 TypeScript sites
- PF-017 reproduced: spread env minus one key
- Comment created with fix

**2. `src/cli/commands/agents.ts:103`** — validateSetArgs missing charset floor (92% confidence, HIGH)  
- Documented trust boundary never wired at CLI entry point
- Cache-miss branch accepts ANY string without MODEL_NAME_RE validation
- Fresh installs vulnerable to persistent corruption
- Comment created with fix

**3. `src/cli/agents-view/render.ts:114`** — TUI sanitization gap (90% confidence, HIGH)
- S2 escape-stripping applied to `--list` but not TUI cell renderer
- padToVisible/truncateVisible compute width but return raw unmodified string
- Comment created with fix

**4. `src/core/agent-models.ts:356`** — invalid-model error routed incorrectly (90% confidence, HIGH)
- Error variant added to union but consumer still folds into generic "malformed frontmatter"
- User blames shipped agent file instead of agent-models.json
- Bad value persists across re-init
- Comment created with switch statement fix

### Medium-Severity Findings

**5. `src/core/model-discovery.ts:743`** — MAX_TTL_MS hardcoded (95% confidence, MEDIUM)
- **Flagged by 5 reviewers across 5 expertise areas** — highest confidence finding
- Constant already exported from cache.ts:36; line 29 already imports that module
- Silent divergence when MAX_TTL_MS changes: readCache() vs getExternalModelsCached() disagree on freshness
- One-line fix: add MAX_TTL_MS to existing import
- Comment created

**6. `src/core/cache.ts:87`** — Dual envelope parsers with divergent strictness (85% confidence, MEDIUM)
- parseEnvelope (private) and parseRawEnvelope (exported) parse same format but diverge on ttl handling
- "Avoid N separate implementations" stated in JSDoc, yet 2 implementations have drifted
- Comment created

**7. `src/cli/commands/proxy.ts:965`** — resolvePort JSDoc orphaned (95% confidence, MEDIUM)
- Two formatters inserted between JSDoc and function (ADR-003 residue)
- JSDoc with TS-1 rationale now misattributed to wrong function
- Section header mislabels contents
- Comment created with structural fix

### Consolidation & Deduplication

Consolidatedoverlapping findings from multiple independent reviewers:
- **ensure-proxy:246**: Raised independently by Security (HIGH, 95%) and Compliance (HIGH, 92%)
- **validateSetArgs**: Raised by Security (MEDIUM, 92%), TypeScript (HIGH, 92%), Regression (HIGH, 90%)
- **invalid-model error**: Raised by TypeScript (HIGH, 90%) and Regression (MEDIUM, 90%)
- **MAX_TTL_MS**: Raised by Consistency (95%), Architecture (95%), Complexity (97%), Reliability (92%), plus Consistency again (95%)
- **TUI sanitization**: Raised by Security (88%) and Consistency (90%)

### Lower-Confidence Findings (60-79%, summary only)

The following findings were referenced in reviews but fell below the 80% threshold or were judgment calls:
- Alternative fixes for validateSetArgs (proactive validation vs narrower provisional path)
- Cache-directory scan triplication (88% arch, 94% complexity) — architectural concern but complex fix
- runEnable rollback triplicated (96% complexity) — worth prioritizing but more complex
- proxy.json convergence on failed init (multiple coverage gaps)
- Charset test for authMode and accountId truncation
- off-cycle-pin computation multiplied across 3 sites

These are captured in the individual reviewer reports for reference during fix phases.

### Status

**Comments created**: 7 inline comments on files with ≥80% confidence findings
**Consolidation**: Heavy overlaps identified and merged into single targeted comments
**Next step**: PR author should address blocking findings before merge

dean0x:
### Inline Comment 1: ensure-proxy:246 — Relay spawn missing allowlist

**Location**: `src/assets/scripts/hooks/ensure-proxy:246`
**Severity**: HIGH (95% confidence)
**Category**: Security / Compliance

The three TypeScript spawn sites were converted to an allowlist (5 allowed env vars) in this range, but the hooks relay spawn—the PRIMARY production spawn path per CLAUDE.md—still uses denylist scoping. This is PF-017 reproduced verbatim.

**Failure scenario**: Developer with OPENAI_API_KEY, GITHUB_TOKEN, SSH_AUTH_SOCK in their shell starts Claude Code. ensure-proxy spawns relay with all inherited. The relay reads 3 env vars; every other secret and signing socket reaches a subprocess with no use for it.

**Fix**:
```bash
set -- PATH="$PATH" HOME="$HOME" SUBSWITCH_CONFIG="$PROXY_CONFIG"
if [ -n "${TMPDIR:-}" ]; then set -- "$@" TMPDIR="$TMPDIR"; fi
if [ -n "${LANG:-}" ];   then set -- "$@" LANG="$LANG"; fi
if [ -n "${LC_ALL:-}" ]; then set -- "$@" LC_ALL="$LC_ALL"; fi
nohup env -i "$@" "$NODE_BIN" "$PROXY_BIN" serve </dev/null >>"$LOG_FILE" 2>&1 &
```

Add regression test to assert stub relay does not observe canary env vars.

dean0x:
### Inline Comment 2: agents.ts:103 — validateSetArgs missing charset floor

**Location**: `src/cli/commands/agents.ts:103-119`
**Severity**: HIGH (92% confidence)
**Category**: Security / TypeScript / Regression

The module documents `MODEL_NAME_RE` as: *"The same charset is used at the CLI entry point so injection payloads are rejected before they can propagate"* (agent-frontmatter.ts:31-32). But grep confirms `isValidModelName` has only one call site—inside `rewriteAgentFrontmatter`. The CLI entry point never calls it.

The cache-miss branch (`catalog.known === false`, the default on fresh installs) accepts **any string** with no validation. This persists arbitrary bytes into `agent-models.json`.

**Failure scenario**: `devflow agents --set coder --model $'sonnet\ntools: Bash'` on fresh install is accepted and persisted. Later reapplyAgentMapping fails with "malformed frontmatter in coder.md (invalid-model)" but points at the wrong file. The poisoned entry persists across re-init and prevents coder from converging.

**Fix**:
```typescript
if (model !== undefined) {
  if (model !== 'default' && !isValidModelName(model)) {
    return Err(`Invalid model name... Allowed: letters, digits, dot, underscore, hyphen.`);
  }
  if (catalog.known) { /* existing allowlist check */ }
}
```

dean0x:
### Inline Comment 3: model-discovery.ts:743 — MAX_TTL_MS hardcoded

**Location**: `src/core/model-discovery.ts:743`
**Severity**: MEDIUM (95% confidence across 5 reviewers)
**Category**: Architecture / Complexity / Consistency / Reliability / TypeScript

`cache.ts:36` exports `MAX_TTL_MS` as the single source of truth, and `model-discovery.ts:29` already imports from that module. Yet line 743 hardcodes the value with a comment that admits the duplication: `7 * 24 * 60 * 60 * 1_000 // mirrors MAX_TTL_MS`

A comment is not a binding. When `MAX_TTL_MS` changes, `readCache()` and `getExternalModelsCached()` silently disagree—one path reports 'cache', the other 'stale-cache'. This was the whole point of exporting the constant.

**Fix** (one line):
```typescript
import { readCache, writeCache, parseRawEnvelope, MAX_TTL_MS } from './cache.js';
// ...
const ttlClamped = Math.min(Math.abs(best.ttl), MAX_TTL_MS);
```

This finding was independently flagged by 5 reviewers—highest cross-discipline confidence in the entire review.

dean0x:
### Inline Comment 4: render.ts:114 — TUI escape-sequence sanitization gap

**Location**: `src/cli/agents-view/render.ts:114-124` + `src/cli/commands/proxy.ts:1011`
**Severity**: HIGH (90% consistency, 88% security)
**Category**: Security / Consistency

The S2 hardening applied `stripAnsi` to the `--list` table (agents.ts:283-286), but the TUI surface renders the same model names raw. Both read the same untrusted `agent-models.json` data.

In render.ts, `padToVisible`/`truncateVisible` call `stripAnsi` **only to compute width** and return the original string unmodified. So unstripped bytes reach the terminal. This defeats the stated purpose: "prevent injection via model IDs".

In proxy.ts:1011, `state.authMode` (from untrusted `~/.codex/auth.json`) is interpolated straight into status output with no sanitization, while its sibling field `accountSuffix` is carefully minimized to 6 chars.

**Fixes**:
- render.ts: sanitize model names before embedding
- proxy.ts: constrain authMode to documented set ('chatgpt'|'apikey'|'unknown') and sanitize

dean0x:
### Inline Comment 5: agent-models.ts:356 — invalid-model error routed to wrong file

**Location**: `src/core/agent-models.ts:356-358`
**Severity**: HIGH (90% confidence)
**Category**: TypeScript / Regression

The `FrontmatterError` union was widened to `'invalid-model'` in this PR, but the sole consumer still folds all variants into one message:

```
reapplyAgentMapping: malformed frontmatter in coder.md (invalid-model) — skipping
```

A user who typos a model name on a fresh install sees the shipped agent file blamed, not the `agent-models.json` entry (the actual problem). The poisoned value persists and re-emits the same misleading warning on every re-init.

**Fix** (add switch for discriminated errors):
```typescript
if (!rewriteResult.ok) {
  switch (rewriteResult.error) {
    case 'invalid-model':
      localWarn(`invalid model name for ${agentName} — fix agent-models.json`);
      break;
    case 'no-frontmatter':
    case 'unterminated-frontmatter':
      localWarn(`malformed frontmatter in ${agentName}.md...`);
      break;
    default: { const _: never = rewriteResult.error; }
  }
  return { bucket: 'skipped', localWarnings };
}
```

dean0x:
### Inline Comment 6: cache.ts — dual envelope parsers diverging

**Location**: `src/core/cache.ts:87` and `:172`
**Severity**: MEDIUM (85% confidence)
**Category**: Architecture / Consistency

`parseEnvelope` (private, line 87) and `parseRawEnvelope` (exported, line 172) both parse the on-disk envelope format but have diverged on the same field:

| Input | `parseEnvelope` | `parseRawEnvelope` |
|-------|---|---|
| non-finite `ttl` | reject entry (→ `null`) | default to 0, accept |

The modules own JSDoc says this function exists *"so model-discovery.ts can share one envelope parser instead of four separate inline re-implementations"* — yet the module retains two parsers that have already drifted.

**Fix**: express one in terms of the other so validation rules are unified.

dean0x:
### Inline Comment 7: proxy.ts:965 — resolvePort JSDoc orphaned

**Location**: `src/cli/commands/proxy.ts:963-1026`
**Severity**: MEDIUM (95% confidence)
**Category**: Architecture / Complexity / Consistency

Two status formatters were inserted in this range **between** `resolvePort`'s doc comment (line 965) and the function itself (line 1026). Result:

- JSDoc block with TS-1 remembered-port rationale is now associated with unrelated `formatExternalModelsLine`
- `resolvePort` itself is now undocumented
- Section header `// ─── Port resolution ───` now mislabels its contents

This is ADR-003 residue: transition scaffolding left in the end-state.

**Fix**: Move `resolvePort`'s JSDoc and section header down to immediately precede the function at line 1026. Give the formatters their own header.

